PR #3, opened 2024-06-07
feat: adding subject template registry

zeroknots:
It might be advantageous to store the subject templates in an external contract.
rhinestone's attestation flow with our registry could hook in here and maintain these subjects.

I also think it would be great, if the templates would be stored per module and not per smart account. making it quite easy for the user to opt-in


src/EmailAccountRecoveryNew.sol
```diff
@@ -177,9 +181,26 @@ abstract contract EmailAccountRecoveryNew {
...
         }
-        for (uint256 idx = 0; idx < recoverySubjectTemplates[account].length; idx++) {
+
+        // TODO: can we get the validator module address from the emailAuthMsg before processing the
+        // entire subject template?
+        // address validatorModule; // TODO
+
+        string[][] memory recoverySubjectTemplate;
+        // get recovery subject template from the RecoverySubjectRegistry
+        if (account == validatorModule) {
```
zeroknots: this would handle the case of the native safe account changeOwner recovery

src/EmailAccountRecoveryNew.sol
```diff
@@ -145,6 +147,8 @@ abstract contract EmailAccountRecoveryNew {
     /// the subject in the given email auth message.
     function handleAcceptance(
         address account,
+        address validatorModule, // idk if this is even possible or if there is a contraint that
```
zeroknots: is this even possible with you current design? im not quite sure if I understand where this function is invoked. So maybe this is a dumb idea.

JohnGuilding: Potentially might be a constraint as this contract was originally designed for non-modular accounts (That is generally the constraint from using this file). This is the function that is called from the relayer after the guardian has replied to their email, confirming they accept to being a guardian.

The issue would be making this interface work for modular and non-modular accounts. Maybe that could be a shared interface, or maybe two different interfaces and the relayer has to figure out which function/contract to call. 

Perhaps one option that is similar but isn't too safe specific would be an `isModularAccount` flag. In this case, the relayer would have to know about the account being modular or not - which I think is possible through an api change.

src/EmailAccountRecoveryNew.sol
JohnGuilding: In order to minimise changes to this file, I'm wondering if this check can be done another way or in another contract.

PR #12, opened 2024-06-30
update-factory-and-split-recovery-module

JohnGuilding:
This PR:
1. Updates the factory to deploy to deterministic addresses. This is to make it compatible with the Rhinestone registry attestation flow.
2. Adds a new module that is less generic, but with stronger security properties. Target validator and function selector are immutable variables
3. Old recovery module renamed to `UniversalEmailRecoveryModule`

The main changes are:
1. `EmailRecoveryFactory.sol`
2. `UniversalEmailRecoveryModule.sol` - same module as before but renamed
3. New `EmailRecoveryModule.sol` - non-generic, but simpler email recovery module with stronger security properties (target validator and function selector passed in at deployment and stored as immutable variables)
4. `EmailRecoveryManager.sol` - modified checks to see if module is authorized to recover


wshino (approved):
LGTM, I will wait @zeroknots approval.

PR #13, opened 2024-07-01
feat: using base module contract to reduce LoC of reused code

zeroknots:
I compartmentalized code that is reused across module flavors into a module base. 
this should reduce LoC for the audit

### DIFF
--- a/src/modules/EmailRecoveryModule.sol
+++ b/src/modules/EmailRecoveryModule.sol
@@ -6,6 +6,7 @@ import { IERC7579Account } from "erc7579/interfaces/IERC7579Account.sol";
 import { IModule } from "erc7579/interfaces/IERC7579Module.sol";
 import { IEmailRecoveryModule } from "../interfaces/IEmailRecoveryModule.sol";
 import { IEmailRecoveryManager } from "../interfaces/IEmailRecoveryManager.sol";
+import { RecoveryModuleBase } from "./RecoveryModuleBase.sol";
 
 /**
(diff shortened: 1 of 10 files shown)


JohnGuilding:
Nice, thanks for trying this out. I'm currently 50/50 on this idea.

On one hand, it does consolidate common logic so it is only written once. On the other hand, I compared these changes to the original module, and LOC are actually a little higher than the current change (guess from the extra boilerplate of base module and not being a huge amount of logic to abstract in terms of LOC in the first place).

@wshino @SoraSuegami wdyt? I think the question I'm trying to answer is is this extra abstraction worth any potential saving from audit cost

wshino:
@JohnGuilding If we plan to create various modules in the future by inheriting from the base module, then this change makes sense.
Are there any real plans to add more modules in the near future? Sorry I am not sure about the future plans.

Divide-By-0:
Less code duplication and cleaner code is good. Don't work about audit cost increases, it's not really substantial or worth factoring in. I would much rather have better code than an easier audit. On the other hand, increasing simplicity is a fine reason to reduce LoC but that doesn't seem to be the case here.

JohnGuilding:
I'm leaning towards keeping things as they are. 

The base module does increases code reuse, but leads to more LOC (extra boilerplate from base module not being offset from reused logic) and increased abstraction. And IMO we should avoid inheritance if it doesn't have a strong benefit

Let me know if anyone disagrees

PR #14, opened 2024-07-01
Factory Improvements

zeroknots:
My understanding is that the verifier, DKIM registry, and email authentication implementation are contracts with security significance. Allowing the caller of the factory to choose these contracts would result in ambiguous security assumptions for monolithic deployments created through the factory.
I recommend making the DKIM registry, verifier, and email authentication implementation immutable to ensure that all managers deployed from the factory adhere to consistent security assumptions.
Moreover, I propose the addition of a functionality for the future-proof deployment of a manager for a recovery module implementation. This feature would not rely on implicit security assumptions but would instead prepare the setup and the factory for future versions or third-party recovery module integrations.

### DIFF
--- a/src/EmailRecoveryFactoryEasy.sol
+++ b/src/EmailRecoveryFactoryEasy.sol
@@ -1,0 +1,137 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.25;
+
+import { Create2 } from "@openzeppelin/contracts/utils/Create2.sol";
+import { EmailRecoveryManager } from "./EmailRecoveryManager.sol";
+import { UniversalEmailRecoveryModule } from "./modules/UniversalEmailRecoveryModule.sol";
+import { EmailRecoveryModule } from "./modules/EmailRecoveryModule.sol";
+
+contract EmailRecoveryFactoryEasy {
+    address public immutable VERIFIER;
+    address public immutable DKIM_REGISTRY;
+    address public immutable EMAIL_AUTH_IMPL;
+
+    event ZkEmailRecoveryModuleCreated(
+        address emailRecoveryModule, address emailRecoveryManager, address subjectHandler
+    );
+
+    event ZkEmailRecoveryManager(
+        address emailRecoveryManager, address subjectHandler, address forModule
+    );
+
+    constructor(address verifier, address dkimRegistry, address emailAuthImpl) {
+        VERIFIER = verifier;
+        DKIM_REGISTRY = dkimRegistry;
+        EMAIL_AUTH_IMPL = emailAuthImpl;
+    }
+
+    function deployForModule(
+        address module,
+        bytes calldata subjectHandlerBytecode,
+        bytes32 subjectHandlerSalt,
+        bytes32 recoveryManagerSalt
+    )
+        external
+        returns (address emailRecoveryManager, address subjectHandler)
+    {
+        // Deploy subject handler
+        subjectHandler = Create2.deploy(0, subjectHandlerSalt, subjectHandlerBytecode);
+        // Deploy recovery manager
+        emailRecoveryManager = address(
+            new EmailRecoveryManager{ salt: recoveryManagerSalt }(
+                VERIFIER, DKIM_REGISTRY, EMAIL_AUTH_IMPL, subjectHandler
+            )
+        );
+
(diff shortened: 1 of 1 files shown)


src/EmailRecoveryFactoryEasy.sol
```diff
@@ -0,0 +1,137 @@
...
+    );
+
+    constructor(address verifier, address dkimRegistry, address emailAuthImpl) {
+        VERIFIER = verifier;
+        DKIM_REGISTRY = dkimRegistry;
+        EMAIL_AUTH_IMPL = emailAuthImpl;
+    }
+
+    function deployForModule(
+        address module,
```
zeroknots, line 29: a counterfactual address could be provided here.

src/EmailRecoveryFactoryEasy.sol
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    function predictAddressesForModule(
+        bytes32 subjectHandlerSalt,
+        bytes calldata subjectHandlerBytecode,
+        bytes32 recoveryManagerSalt
+    )
+        external
+        view
+        returns (address subjectHandler, address emailRecoveryManager)
```
zeroknots, line 59: `emailRecoveryManager` address would be used in the constructor of the 3rd party recovery module

src/EmailRecoveryFactoryEasy.sol
```diff
@@ -0,0 +1,137 @@
...
+            keccak256(
+                abi.encodePacked(
+                    type(EmailRecoveryManager).creationCode,
+                    abi.encode(VERIFIER, DKIM_REGISTRY, EMAIL_AUTH_IMPL, subjectHandler)
+                )
+            )
+        );
+    }
+
+    function createZkEmailRecoveryModule(
```
zeroknots, line 75: rewrote these functions

src/EmailRecoveryFactoryEasy.sol
```diff
@@ -0,0 +1,137 @@
...
+        bytes32 subjectHandlerSalt,
+        bytes32 recoveryManagerSalt,
```
zeroknots, line 77: im actually not sure if create2 is a huge upside here.

src/EmailRecoveryFactoryEasy.sol
```diff
@@ -0,0 +1,137 @@
...
+        address module,
+        bytes calldata subjectHandlerBytecode,
+        bytes32 subjectHandlerSalt,
+        bytes32 recoveryManagerSalt
+    )
+        external
+        returns (address emailRecoveryManager, address subjectHandler)
+    {
+        // Deploy subject handler
+        subjectHandler = Create2.deploy(0, subjectHandlerSalt, subjectHandlerBytecode);
```
zeroknots, line 38: create2 necessary here to allow counterfactual deployments

JohnGuilding:
> I recommend making the DKIM registry, verifier, and email authentication implementation immutable to ensure that all managers deployed from the factory adhere to consistent security assumptions.

Agreed, will make this change. Will check if the dkim registry should still be passed in, in the case of a user specified dkim registry.

> Moreover, I propose the addition of a functionality for the future-proof deployment of a manager for a recovery module implementation. This feature would not rely on implicit security assumptions but would instead prepare the setup and the factory for future versions or third-party recovery module integrations.

I'm a little uncertain on this idea. Can you steel-man the argument for it?

I think my apprehension (which admittedly is not fully fleshed out) is from:
1. uncertainty on how this will work in the attestation flow. The question in my mind is will separating the recovery module and manager/handler deployment have any negative consequences?
2. I do like how the deployment of the module, manager and subject handler is tightly coupled in these factory functions. Although maybe this isn't a valid benefit as maybe deploying them separately works well with the attestation flow and we can confidently attest to the combined security of the different contract combinations

src/EmailRecoveryFactoryEasy.sol
JohnGuilding: To clarify, which create2?

JohnGuilding:
Resolved by https://github.com/zkemail/email-recovery/pull/16. 

Raised a separate PR as wanted to merge quickly and only take parts of this PR.

@zeroknots thanks for raising these changes and still interested to hear your thoughts on having the additional functions for future proofing :)

PR #15, opened 2024-07-01
Modify check for no guardians & add tests

JohnGuilding:
Builds upon changes slightly from https://github.com/zkemail/email-recovery/commit/ca7d21c9528e1cb7df8dedd6da0458688387233d and adds tests

1. Only calls `setupGuardians` if guardians.length, weights.length, and threshold are all 0. It felt appropriate that no guardian weights should be passed in if the guardians themselves are not known. And passing the threshold also felt like it could be redundant without knowledge of guardian count and weights. The logic added in commit https://github.com/zkemail/email-recovery/commit/ca7d21c9528e1cb7df8dedd6da0458688387233d failed when running a test with no guardians passed in because of conflicting checks in `setupGuardians`. Permitting weight and threshold also be zero and adding the check in `configureRecovery` keeps the `setupGuardian` function simple.

4. Adds `guardianConfig.initialized` checks in existing tests
5. Adds tests for new logic in configureRecovery.t.sol

### DIFF
--- a/src/libraries/GuardianUtils.sol
+++ b/src/libraries/GuardianUtils.sol
@@ -49,7 +49,7 @@ library GuardianUtils {
             revert IncorrectNumberOfWeights();
         }
 
-        if (threshold == 0 && guardianCount > 0) {
+        if (threshold == 0) {
             revert ThresholdCannotBeZero();
         }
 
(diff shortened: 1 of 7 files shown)


src/libraries/GuardianUtils.sol
```diff
@@ -49,7 +49,7 @@ library GuardianUtils {
             revert IncorrectNumberOfWeights();
         }
 
-        if (threshold == 0 && guardianCount > 0) {
```
SoraSuegami, line 52: You removed `&& guardianCount > 0` because `setupGuardians` will be never called in `configureRecovery` when no guardian is set, right?

SoraSuegami (approved):

src/libraries/GuardianUtils.sol
JohnGuilding: Yes exactly

PR #20, opened 2024-07-04
chore: Update callling function name in DeployEmailRecoveryModuleScript.

wshino:
Clave wallet needs `resetOwners(bytes)` calling not `changeOwner(address)`.


SoraSuegami:
Could you take the selector name from env and set its default value `changeOwner(address)`?
As the script is used for various wallets, I think it is better to manually put `resetOwners(bytes)` as the env for clave wallets than hardcoding it.

JohnGuilding:
This change is not needed

PR #22, opened 2024-07-18
Ackee audit fixes

JohnGuilding:
### Notes:

**H2: Premature guardian configuration update in addGuardian function**
Decided to add `acceptedWeight` as a dedicated variable, as the alternative was to loop over guardians every time processRecovery is called, which could be quite computationally expensive. Adding the dedicated variable didn’t actually add too much complexity

**M3: Selector collisions in UniversalEmailRecoveryModule**
The simplest solution to this was to remove the `selectorToValidator` mapping and just pass the validator in with the calldata to `recover`. I used assembly to retrieve the selector as the `recoveryCalldata` is now stored in memory and not in calldata since it is abi decoded along with the target validator

**W4: Gas inefficiencies in UniversalRecoveryModule**
Issue 1 was fixed by the function selector collision change
Issue 2 is resolved
Issue 3 - `getAllowedValidators` needed to be called anyway so didn’t implement - unless you see a way it isn’t needed we missed

**W6: Missing AddedGuardian event emission in setupGuardians function**
Emitted the AddedGuardian event and simplified the code by just calling `addGuardian` inside `setupGuardians`

**W7: ERC-4337 violation in onInstall**
This was a great catch. Resolved by adding comments explaining the violation as decided it was too complex to make it compatible. Future versions could look to resolve this

**I6: Floating pragma**
Chose not to implement for better compatibility with external contracts

**I9: Safe validateRecoverySubject optimization**
Didn’t need to combine with `getPreviousOwnerInLinkedList` as that was moved into the `parseRecoveryCalldata` function. Did check the `newOwner` against existing owners


src/EmailRecoveryManager.sol
```diff
@@ -225,8 +234,8 @@ contract EmailRecoveryManager is EmailAccountRecovery, Initializable, IEmailReco
      * @param expiry The expiry time after which the recovery attempt is invalid
      */
     function configureRecovery(
-        address[] memory guardians,
-        uint256[] memory weights,
+        address[] calldata guardians,
```
wshino, line 237: If there are no changes for guardians, calldata is more appropriate than memory. Is that the reason this change?

src/interfaces/IEmailRecoveryManager.sol
```diff
@@ -39,20 +39,33 @@ interface IEmailRecoveryManager {
...
      */
     struct GuardianConfig {
-        uint256 guardianCount;
-        uint256 totalWeight;
-        uint256 threshold;
-        bool initialized;
+        uint256 guardianCount; // total count for all guardians
+        uint256 totalWeight; // combined weight for all guardians. Important for checking that
+            // thresholds are valid.
+        uint256 acceptedWeight; // combined weight for all accepted guardians. This is separated
```
wshino, line 45: Nice

src/EmailRecoveryManager.sol
```diff
@@ -380,16 +388,18 @@ contract EmailRecoveryManager is EmailAccountRecovery, Initializable, IEmailReco
 
         recoveryRequest.currentWeight += guardianStorage.weight;
 
-        uint256 threshold = guardianConfigs[account].threshold;
-        if (recoveryRequest.currentWeight >= threshold) {
+        if (recoveryRequest.currentWeight >= guardianConfig.threshold) {
+            bytes32 calldataHash = IEmailRecoverySubjectHandler(subjectHandler)
+                .parseRecoveryCalldataHash(templateIdx, subjectParams);
```
wshino, line 393: Was this function separated from validateRecoverySubject to return calldataHash efficiently?

src/modules/EmailRecoveryModule.sol
```diff
@@ -43,16 +44,28 @@ contract EmailRecoveryModule is ERC7579ExecutorBase, IEmailRecoveryModule {
...
+        if (_emailRecoveryManager == address(0)) {
+            revert InvalidManager();
+        }
+        if (_validator == address(0)) {
+            revert InvalidValidator(_validator);
+        }
+        if (
+            _selector == IModule.onUninstall.selector || _selector == IModule.onInstall.selector
+                || _selector == IERC7579Account.execute.selector
+                || _selector == ISafe.setFallbackHandler.selector
```
wshino, line 66: Could you tell me why we need to check FallbackHandler.selector?

wshino (commented):
I left some comments including questions.

src/EmailRecoveryManager.sol
JohnGuilding: yeah, that is why they are now calldata?

Although I change them back in the consolidation PR as the function is called directly by the module. As they are abi decoded in `onInstall`, this means they are in memory so these params have to be changed back to memory. That is for the other PR though

src/EmailRecoveryManager.sol
JohnGuilding: Yeah it was a recommendation from the auditors to reduce unnecessary computation every time `processRecovery` was called

src/modules/EmailRecoveryModule.sol
JohnGuilding: It was a recommendation from the auditors. A 7579 Safe can be used as a validator. And that contract can execute arbitrary logic on a safe including the function - `setFallbackHandler`. Adding this check stops that function being called directly from the Safe on recovery as it could cause some damage if it was called

wshino (approved):
Looks good, thank you so much for your changes.

PR #24, opened 2024-07-30
Feat/native safe recovery module

JohnGuilding:
* Add native Safe module
* Make `configureRecovery` function in manager `public` instead of `internal` so it can be called for the native Safe module. This is because Safe modules do not have an equivalent to the `onInstall` function
* Add native Safe recovery module integration tests

### DIFF
--- a/script/DeploySafeNativeRecovery.s.sol
+++ b/script/DeploySafeNativeRecovery.s.sol
@@ -4,36 +4,18 @@ pragma solidity ^0.8.25;
 import { Script } from "forge-std/Script.sol";
 import { console } from "forge-std/console.sol";
 import { SafeEmailRecoveryModule } from "src/modules/SafeEmailRecoveryModule.sol";
-import { Verifier } from "ether-email-auth/packages/contracts/src/utils/Verifier.sol";
-import { ECDSAOwnedDKIMRegistry } from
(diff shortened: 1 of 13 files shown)


script/DeploySafeNativeRecovery.s.sol
```diff
@@ -0,0 +1,24 @@
...
+import { console } from "forge-std/console.sol";
+import { SafeEmailRecoveryModule } from "src/modules/SafeEmailRecoveryModule.sol";
+
+contract DeploySafeNativeRecovery_Script is Script {
+    function run() public {
+        vm.startBroadcast(vm.envUint("PRIVATE_KEY"));
+        address manager = vm.envOr("VERIFIER", address(0));
+
+        if (manager == address(0)) {
+            manager = address(1);
```
wshino: Is this address(1) as a placeholder to run this script?
Is it okay if we don't revert?

wshino (commented):
one comment.

script/DeploySafeNativeRecovery.s.sol
JohnGuilding: Yeah this is just a placeholder, I need to update this PR to the latest changes from main

wshino (approved):
LGTM

PR #26, opened 2024-07-31
Consolidate module & manager into single contract

JohnGuilding:

PR #28, opened 2024-08-04
Add functions which returns the user state is in module enabled or recovery is activated.

wshino:
This PR is for adding the functions which returns user state is in module enabled or recovery process is activated.

~~Currently for rhinestone, they removed the branch named feature/4337-compliance in rhinestonewtf/safe7579.
I used forking repo in zkemail instead, I hope this will be fixed soon. 
I'm going to create a PR to them.~~
Also regarding ether-email-auth, at now we need to use commit hash or branch name to use the latest EmailAccountRecovery.sol. This will be fixed soon as we will use npm package.

### DIFF
--- a/src/EmailRecoveryManager.sol
+++ b/src/EmailRecoveryManager.sol
@@ -105,13 +105,12 @@ abstract contract EmailRecoveryManager is
 
     /**
      * @notice Checks if the recovery is activated for a given account
-     * @param recoveredAccount The address of the recovered account for which the activation status is being checked
+     * @param account The address of the account for which the activation status is being checked
      * @return bool True if the recovery request is activated, false otherwise
      */     
-    function isActivated(address recoveredAccount) public view override returns (bool) {
-        return guardianConfigs[recoveredAccount].threshold > 0;
+    function isActivated(address account) public view override returns (bool) {
+        return guardianConfigs[account].threshold > 0;
     }
-
 
     /**
      * @notice Returns a two-dimensional array of strings representing the subject templates for an
@@ -283,6 +282,10 @@ abstract contract EmailRecoveryManager is
             revert RecoveryInProcess();
         }
 
+        if (!isActivated(account)) {
+            revert RecoveryIsNotActivated();
+        }
+
         // This check ensures GuardianStatus is correct and also implicitly that the
         // account in email is a valid account
         GuardianStorage memory guardianStorage = getGuardian(account, guardian);
@@ -322,6 +325,10 @@ abstract contract EmailRecoveryManager is
         address account = IEmailRecoverySubjectHandler(subjectHandler).validateRecoverySubject(
             templateIdx, subjectParams, address(this)
         );
+
+        if (!isActivated(account)) {
+            revert RecoveryIsNotActivated();
+        }
 
         GuardianConfig memory guardianConfig = guardianConfigs[account];
         if (guardianConfig.threshold > guardianConfig.acceptedWeight) {
@@ -378,6 +385,10 @@ abstract contract EmailRecoveryManager is
         }
         RecoveryRequest memory recoveryRequest = recoveryRequests[account];
 
+        if (!isActivated(account)) {
+            revert RecoveryIsNotActivated();
+        }
+        
         uint256 threshold = guardianConfigs[account].threshold;
         if (threshold == 0) {
             revert NoRecoveryConfigured();
--- a/src/EmailRecoveryManager.sol
+++ b/src/EmailRecoveryManager.sol
@@ -107,7 +107,7 @@ abstract contract EmailRecoveryManager is
      * @notice Checks if the recovery is activated for a given account
      * @param account The address of the account for which the activation status is being checked
      * @return bool True if the recovery request is activated, false otherwise
-     */     
+     */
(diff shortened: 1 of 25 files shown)


src/modules/UniversalEmailRecoveryModule.sol
```diff
@@ -229,6 +229,15 @@ contract UniversalEmailRecoveryModule is
         return getGuardianConfig(account).threshold != 0;
     }
 
+    /**
+     * Check if the recovery module is authorized to recover the account
+     * @param account The smart account to check
+     * @return true if the module is authorized, false otherwise
+     */
+    function isAuthorizedToBeRecovered(address account) external view returns (bool) {
+        return validatorCount[account] > 0;
```
wshino: Should I use `return authorized[account]` as same as the src/modules/EmailRecoveryModule.sol ?

src/EmailRecoveryManager.sol
```diff
@@ -103,6 +103,16 @@ abstract contract EmailRecoveryManager is
         return recoveryRequests[account];
     }
 
+    /**
+     * @notice Checks if the recovery is activated for a given account
+     * @param recoveredAccount The address of the recovered account for which the activation status is being checked
+     * @return bool True if the recovery request is activated, false otherwise
+     */     
+    function isActivated(address recoveredAccount) public view override returns (bool) {
+        return guardianConfigs[recoveredAccount].threshold > 0;
```
SoraSuegami, line 112: I think this function returns false even after the account installs the module because they can install it without setting any guardians as the argument of `onInstall` function.

src/modules/EmailRecoveryModule.sol
```diff
@@ -34,12 +34,17 @@ contract EmailRecoveryModule is EmailRecoveryManager, ERC7579ExecutorBase, IEmai
      */
     bytes4 public immutable selector;
 
+    /**
+     * Mapping of Account address to authorized validator
+     */
+    mapping(address account => bool isAuthorized) internal authorized;
```
SoraSuegami, line 40: Sorry, I don't understand why you need to introduce this new mapping.

SoraSuegami (commented):
Thank you for your PR!
I will check codes.

src/EmailRecoveryManager.sol
```diff
@@ -103,6 +103,16 @@ abstract contract EmailRecoveryManager is
...
+    }
+
+
```
JohnGuilding:
```suggestion
    /**
     * @notice Checks if the recovery is activated for a given account
     * @param account The address of the account for which the activation status is being checked
     * @return bool True if the recovery request is activated, false otherwise
     */     
    function isActivated(address account) public view override returns (bool) {
        return guardianConfigs[account].threshold > 0;
    }
```

JohnGuilding: Nitpick - added suggestion for consistent naming with other functions in this file

test/unit/EmailRecoveryManager/updateRecoveryConfig.t.sol
JohnGuilding, line 1: Nitpick - why do we need assertions for `isActivated ` in these tests? `threshold` state isn't modified in this function

src/modules/UniversalEmailRecoveryModule.sol
JohnGuilding: See comment on modified solution. Let me know what you think

JohnGuilding (changes requested):
This looks good. Although I think I've thought of a better way of doing this. Wdyt of this idea:

Instead of re-adding the `isAuthorizedToBeRecovered` logic, we remove that function and related state from each recovery module, and just use the `isActivated` function as you use it here? There is no need to write a separate function on the modules if you can use `isActivated` and you can use state from the manager, as the manager is now part of the module

If you agree with this change, could you also re-add the logic for calling `isAuthorizedToBeRecovered` in the manager too but using `isActiviated` instead? Like this:
```solidity
if (!isActivated()) {
    revert RecoveryModuleNotAuthorized();
}
```
It isn't strictly related to the `isActiviated` change, but will actually be useful for the native safe module PR (it was my mistake to remove it in hindsight). This will involve:
1. Adding the equivalent of [this logic](https://github.com/zkemail/email-recovery/pull/26/files#diff-e8352f426d30b4e5e897fbc510d7cec8c318b41fa9bc66f09b022af29e83b2ccL372-L375), but using `!isActivated()` instead of `isAuthorizedToBeRecovered` back to `acceptGuardian` and `processRecovery`:
```solidity
if (!isActivated()) {
    revert RecoveryModuleNotAuthorized();
}
```
2. Add the tests back for this
3. Also add a similar check in `completeRecovery`:
4. Write a test for new check in `completeRecovery`
5. Remove all changes related to `isAuthorizedToBeRecovered` and associated state in the modules that were added in this PR

JohnGuilding (approved):

JohnGuilding:
LGTM @wshino. Pushed some minor changes:
1. My suggestion to add the `isActivated` check in `completeRecovery` wasn't needed as there is already an equivalent check
2. Added new error selector to `assertErrorSelectors.t.sol`
3. Added unit test for `isActivated`

PR #43, opened 2024-09-03
Ignore validator address in EmailRecoveryModule

JohnGuilding:


src/modules/EmailRecoveryModule.sol
```diff
@@ -148,14 +147,11 @@ contract EmailRecoveryModule is EmailRecoveryManager, ERC7579ExecutorBase, IEmai
      * being recovered. recoveryData = abi.encode(validator, recoveryFunctionCalldata)
      */
     function recover(address account, bytes calldata recoveryData) internal override {
-        (address validator, bytes memory recoveryCalldata) =
+        (, bytes memory recoveryCalldata) =
             abi.decode(recoveryData, (address, bytes));
```
SoraSuegami, line 151: If we don't use an address in `recoveryData`, is it possible/easy to simply remove `address` from the encoding definition of  `recoveryData`?

PR #46, opened 2024-09-18
Feat/test improvements

JohnGuilding:
This PR:
* Extracts duplicate logic from tests into fewer base tests. As opposed from having many base tests with the same logic
* Renames some stuff
* Fixes a load of compile warnings

Might be a difficult one to review without good context on how the tests are structured. Would try:
1. Verify the assertions haven't fundamentally changed. Most of the changes in the test files themselves are just renames or some sort/changing parameters passed into test helper functions
2. Focus on Base test files.
    - That `Base.t.sol` has everything you would expect
    - Review use of `computeEmailAuthAddress` as that is a virtual function in `Base.t.sol` and is implemented in other base tests
    - Review use of `deployModule` as that is a virtual function in `Base.t.sol` and is implemented in other base tests

There is still some work to be done here, for example abstracting away `handleRecovery` helper functions into fewer base test files. That will be covered in a future PR that can be merged on top of the audit fixes


.solhint.json
```diff
@@ -14,6 +14,7 @@
     "not-rely-on-time": "off",
     "one-contract-per-file": "off",
     "var-name-mixedcase": "off",
-    "immutable-vars-naming": "off"
+    "immutable-vars-naming": "off",
+    "no-console": "warn"
```
wshino, line 18: Thank you, I sometimes forget to remove the console.log

test/integration/EmailRecoveryManager/EmailRecoveryManager.integration.t.sol
```diff
@@ -25,35 +23,38 @@ contract EmailRecoveryManager_Integration_Test is
     }
 
     function test_RevertWhen_HandleAcceptanceCalled_BeforeConfigureRecovery() public {
```
wshino, line 25: Is it okay to remove this vm.prank?

wshino (approved):
reviewed and approved.

wshino (approved):
reviewed and approved.

test/integration/EmailRecoveryManager/EmailRecoveryManager.integration.t.sol
JohnGuilding: Yeah it is unnecessary when calling `instance1.uninstallModule` and similar functions on `instance1`

PR #52, opened 2024-10-12
Add script for account hiding

wshino:


JohnGuilding (approved):
We should add a test for this but looks good

PR #55, opened 2024-10-21
Use salt value for deployment

wshino:


SoraSuegami:
Could you add these salts to all scripts to deploy modules?

PR #58, opened 2024-11-01
Add kill switch functionality

JohnGuilding:


src/EmailRecoveryManager.sol
```diff
@@ -577,7 +608,7 @@ abstract contract EmailRecoveryManager is
      * unexpected behaviour when reinstalling account modules, the contract state should be
      * deinitialized. This should include removing state accociated with an account.
      */
-    function deInitRecoveryModule() internal onlyWhenNotRecovering {
+    function deInitRecoveryModule() internal onlyWhenNotRecovering onlyWhenActive {
```
SoraSuegami: I think this function should work even in the emergency case because users will uninstall the module.

JohnGuilding: good catch, agreed

PR #67, opened 2024-12-16
Feat/deterministic deployments

zkfriendly:
Integrates `SafeSingletonDeployer` for deterministic deployments of `UniversalEmailRecoveryModule` across multiple chains.


JohnGuilding (commented):
To make the difference clear between the deployer and the owner, can you do the following
1. define deployer like so:
`deployer = vm.envAddress("DEPLOYER");`
2. Define the owner like so:
`initialOwner = vm.envAddress("INITIAL_OWNER");`
3. could you remove all uses of `KILL_SWITCH_AUTHORIZER` and replace with `INITIAL_OWNER`
4. Basically ensure deployer is the address always deploying, and then the initial owner is always passed into the deploy function as the initial owner. I think the code didn't make this distinction clear before

also build is failing because you removed the override, was that removed because of a compile warning?

zkfriendly:
Fixed the mentioned issues.

> define deployer like so:
>    deployer = vm.envAddress("DEPLOYER");

this cannot be a an address because safe deployer expects a private key rather than an address

> also build is failing because you removed the override, was that removed because of a compile warning?

`ERC7579ValidatorBase` does not have a `validateSignatureWithData` function and I was getting compile *error* running the script saying function has override specified but does not override anything. 

Not sure why the build is failing for the opposite reason here. I'm probably missing something

JohnGuilding (commented):
Would you be able to to do the same as `email-tx-builder` where we modify the underlying base scripts in `BaseDeployScript`?

I will also test this out tomorrow along with `email-tx-builder` changes, then we can get both PRs in

JohnGuilding:
> this cannot be a an address because safe deployer expects a private key rather than an address

I oh see, I thought there was something clever going on under the hood with an address override, where it would default to different options. Private key is explicit and simpler to understand though so sounds good

> Not sure why the build is failing for the opposite reason here. I'm probably missing something

Yeah wonder what was going on here, maybe you had the old version of `ERC7579ValidatorBase` stored in your cache from a previous `pnpm install`?

zkfriendly:
> I oh see, I thought there was something clever going on under the hood with an address override, where it would default to different options. Private key is explicit and simpler to understand though so sounds good

turns out we don't need to explicitly pass the deployer if we use the `deploy` method instead of `deployAndBroadcast` and wrap all deployments in `vm.startBroadcast` and `vm.stopBroadcast`

> Yeah wonder what was going on here, maybe you had the old version of ERC7579ValidatorBase stored in your cache from a previous pnpm install?

That was probably it.

.env.example
```diff
@@ -3,6 +3,9 @@ SEPOLIA_RPC_URL=
 MAINNET_RPC_URL=
 
 PRIVATE_KEY=""
```
JohnGuilding, line 5: Could you remove this and rename DEPLOYER to DEPLOYER_PRIVATE_KEY? And then replace use of PRIVATE_KEY that too

JohnGuilding (dismissed):
Looks good! Only additional comment would be adding foundry version + commit if you're not already testing that out (and it works)

Tested deployments on my side and they behaved as expected

JohnGuilding:
Thanks for doing this work @zkfriendly. Closing this as solving the issues was taking too long and we wanted to deploy to production. We want this in the future though so this was useful to try out.

Issue #66, opened 2024-12-10
Prototype a new ERC7579 recovery module


JohnGuilding:
Implement a prototype for a new ERC7579 recovery module. The new module should be simpler than the existing module and generic to multiple guardian verification methods.

**Requirements:**
* Should be simpler than existing recovery module
* Can reuse existing logic generously, but must not reuse entire contracts. **Hint**: you  will need to inspect and use recovery logic in https://github.com/zkemail/email-tx-builder as well as this repo. This includes implementing `IVerifier.sol` & `IDKIMRegistry.sol`
* Should be used with an additional guardian verification mechanism other than ZK Email. For the prototype, an EOA guardian is sufficient
* Does not have to be feature complete as a prototype, but must be able to recover a 7579 smart account
* Must compile and have a single happy path test

rose2221:
i'd like to contribute to this issue

rose2221:
i'd like to work on this issue!

JohnGuilding:
Hey @rose2221, great

askwhyharsh:
ok, working on this.  

1 - will it take the same params as in the current recovery module?

i.e
   ```
        address verifier,
        address dkimRegistry,
        address emailAuthImpl,
        address commandHandler,
        uint256 minimumDelay,
        address killSwitchAuthorizer,
        address _validator,
        bytes4 _selector

Arch0125:
@JohnGuilding Hey are you guys still open for submissions ?

JohnGuilding:
We've now closed submissions for the solidity engineer role. All those with submissions of sufficient quality have been contacted already for compensation. If you have not been contacted, your submission did not meet the requirements of the github issue

PR #75, opened 2025-01-06
 prototype for a generic ERC7579 recovery module

0xnonso:
see #66 
### _Approach_
This prototype preserves the initial flow of the recovery module with some modifications.
In order to make the recovery module more generic for multiple guardian types, this prototype separates the guardian verification mechanism from the recovery module contract which makes the core recovery module much more simpler. This separation allows for greater flexibility in supporting different kinds of guardians. For example, a user could choose to use a P-256 public key as a guardian, an EOA etc.
### _Smart contract changes_
* The guardian type is changed to `bytes` to support multiple types using `abi.encode(...)`.
* The guardian is not stored directly; instead, it is hashed and then stored in an enumerable map.
* Mapping which stores whether the guardian verifier is supported by an account.
* Mapping of the guardian identifier for an account to its guardian verifier.

Some example verifiers where created, such as `EmailGuardianVerifier.sol` and `EDCSAGuardianVerifier.sol` to demonstrate what the interface for guardian verifiers would look like (`EmailGuardianVerifier.sol` has not been tested; it is included only to illustrate how it could be implemented as a separate verifier).

### _Testing_
The prototype tests demonstrate a happy path for recovering a validator using a direct EOA (msg.sender), ECDSA signatures, and a combination of both.
to run tests: `forge test --match-path 'test/prototype/**/*.t.sol'`

all changes can be found in  `src/prototype/*` and `test/prototype/*`


JohnGuilding:
Thank you for the submission Nonso

PR #76, opened 2025-01-09
Implement ERC7579 Simple Recovery Module Prototype

### LINKED ISSUE #66: Prototype a new ERC7579 recovery module

rose2221:
Closes #66

This PR introduces a prototype for a new, simplified recovery module located at `src/simpleRecoveryPrototype`.  

For testing, run:  
```bash
forge test --match-contract RecoveryTest
```  

The detailed approach has already been shared via mail.  

Thanks!

### DIFF
--- a/src/simpleRecoveryPrototype/EOA712Verifier.sol
+++ b/src/simpleRecoveryPrototype/EOA712Verifier.sol
@@ -1,0 +1,140 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.17;
+
+import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+import "@openzeppelin/contracts/utils/cryptography/SignatureChecker.sol";
+
+
+/**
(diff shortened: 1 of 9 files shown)


JohnGuilding:
Thank you for the submission Rose

PR #80, opened 2025-01-31
feat: add recovery module with tests

### LINKED ISSUE #66: Prototype a new ERC7579 recovery module

Arch0125:
# Modular and almost stateless ERC7579 recovery module 

## Approach
The Recovery module has been designed to increase the modularity and statelessness of the module. Keeping that in mind, the module is stripped off from the current state to be minimalistic to offload the guardian/verification related executions to external contracts, which can be set by the user or whitelisted by the module itself. This allows support for multiple verifiers without upgrading/redeploying the existing module contract.

Hence, the module is totally free of any external logic and can be used according to user's choice

## Modular and Stateless aspects
- The module doesn't acts as the guardian manager itself hence it's modular in the aspect the user can come in with any preferred guardian manager
- The proofs are transient and hence not stored in the module itself
- The verifiers are set in a modular way so that the execution of proofs are done over the external contracts, hence the module is not overloaded with verification logic
- The verifiers can be white-listed by the recovery module

## Future improvements
- Allowing more generalized calls to the guardian manager
- Add support for more verifiers
- Add verifier support for P256 verifier as well

### Running test
```
forge test --match-test test_RecoverAccount_With_EOA_ZKEmail -vvvv
```

Closes #66

### DIFF
--- a/src/SimpleGuardianManager.sol
+++ b/src/SimpleGuardianManager.sol
@@ -1,0 +1,121 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.12;
+
+contract SimpleGuardianManager {
+    struct GuardianConfig {
+        address[] guardians;
+        mapping(address => uint256) weights;
+        uint256 threshold;
+    }
+
+    struct RecoverySession {
+        address[] proofSubmitters;
+        bytes[] proofs;
+        uint256 timestamp;
+        bool active;
+    }
+
+    mapping(address => GuardianConfig) public guardianConfigs;
+    mapping(address => mapping(uint256 => RecoverySession)) private recoverySessions;
+
+    uint256 public constant PROOF_VALIDITY_PERIOD = 3 days;
+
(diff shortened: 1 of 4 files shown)


JohnGuilding:
Thank you for the submission Archisman

PR #81, opened 2025-02-03
added unit tests for transaction initiater changing

aalimsahin:
# WHAT

added the 'transactionInitiator' change to the EmailRecoveryModule contract
added unit tests for changes.

# TEST
To test the changes, please uncomment the commented lines in the test/integration/OwnableValidatorRecovery/EmailRecoveryModule/EmailRecoveryModuleBase.t.sol (lines 123-149) and test/unit/modules/EmailRecoveryModule/EmailRecoveryModuleBase.t.sol (lines 78-102) files. Don't forget to comment out the previous one each time.

0- No changes were made to the old existing tests. We can see the failed tests along with the changes made to the Email Recovery Module contract. (Expect 15 tests to fail under this condition.)

1- After deploying the EmailRecoveryModule, assign the address that will call the acceptGuardian and processRecovery functions as the transactionInitiator. (All tests must be passed successfully)

2- Verify that everyone can initiate transactions after 6 months, even if no transactionInitiator is assigned. (All tests must be passed successfully)

3- Verify that the setTransactionInitiator function works as expected. (Expect 15 tests to fail under this condition.)

4- Validate the behavior when the transaction initiator flag is set to true first, then changed to false, and the contract is working after a waiting period of 6 months. (All tests must be passed successfully)

5- Verify that only the killSwitchAuthorizer can set the transaction initiator. (The setup phase of the tests must be failed.)

-- New --

6- Verify that it does not work before 6 months when the transaction initiator flag is not true. (Expect 15 tests to fail under this condition.)

### DIFF
--- a/src/modules/EmailRecoveryModule.sol
+++ b/src/modules/EmailRecoveryModule.sol
@@ -105,6 +105,9 @@ contract EmailRecoveryModule is EmailRecoveryManager, ERC7579ExecutorBase, IEmai
 
         validator = _validator;
         selector = _selector;
+
+
+        deploymentTimestamp = block.timestamp;
     }
 
     /*´:°•.°+.*•´.*:˚.°*.˚•´.°:°•.°•.*•´.*:˚.°*.˚•´.°:°•.°+.*•´.*:*/
--- a/src/modules/EmailRecoveryModule.sol
+++ b/src/modules/EmailRecoveryModule.sol
@@ -33,7 +33,32 @@ contract EmailRecoveryModule is EmailRecoveryManager, ERC7579ExecutorBase, IEmai
      */
     bytes4 public immutable selector;
 
+    /**
+     * Deployment timestamp
+     */
+    uint256 public immutable deploymentTimestamp;
+
+    /**
+     * Account address to initiate transactions
+     */
+    mapping(address account => bool isInitiator) internal transactionInitiators;
+
     event RecoveryExecuted(address indexed account, address indexed validator);
+
+    /**
+     * @notice Modifier to check if the caller is an initiator
+     */
+    modifier isInitiator() {
+        bool isOpenToAll = transactionInitiators[address(0)]
+            || block.timestamp >= deploymentTimestamp + 6 * 30 days;
+
+        if (!isOpenToAll) {
+            require(
+                transactionInitiators[msg.sender], "Only allowed accounts can call this function"
+            );
+        }
+        _;
+    }
 
     error InvalidSelector(bytes4 selector);
     error InvalidOnInstallData();
@@ -125,6 +150,14 @@ contract EmailRecoveryModule is EmailRecoveryManager, ERC7579ExecutorBase, IEmai
     }
 
     /**
+     * @notice Sets the transaction initiator status for an account
+     * @dev Can only be called by the kill switch authorizer
+     */
+    function setTransactionInitiator(address account, bool canInitiate) external onlyOwner {
+        transactionInitiators[account] = canInitiate;
+    }
+
+    /**
      * @notice Check if the module is initialized
      * @param account The smart account to check
      * @return bool True if the module is initialized, false otherwise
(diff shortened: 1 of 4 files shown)


src/modules/EmailRecoveryModule.sol
```diff
@@ -33,8 +33,33 @@ contract EmailRecoveryModule is EmailRecoveryManager, ERC7579ExecutorBase, IEmai
      */
     bytes4 public immutable selector;
 
+    /**
+     * Deployment timestamp
+     */
+    uint256 public immutable deploymentTimestamp;
```
JohnGuilding, line 39: Not set in constructor

src/modules/EmailRecoveryModule.sol
```diff
@@ -201,4 +234,53 @@ contract EmailRecoveryModule is EmailRecoveryManager, ERC7579ExecutorBase, IEmai
...
+     */
+    function acceptGuardian(
+        address guardian,
+        uint256 templateIdx,
+        bytes[] memory commandParams,
+        bytes32 nullifier
+    )
+        internal
+        override(EmailRecoveryManager)
+        onlyWhenActive
```
JohnGuilding, line 261: Was there a reason `onlyWhenActive` was added here? `onlyWhenActive` will be called twice so can be safely  removed here and for `processRecovery`. Unless you wanted to execute it twice but always show the `KillSwitchEnabled` first if that error occurrs at the same time as `"Only allowed accounts can call this function"`

src/modules/EmailRecoveryModule.sol
```diff
@@ -201,4 +234,53 @@ contract EmailRecoveryModule is EmailRecoveryManager, ERC7579ExecutorBase, IEmai
...
+     * @param nullifier The unique identifier for an email (unused in this implementation)
+     */
+    function acceptGuardian(
+        address guardian,
+        uint256 templateIdx,
+        bytes[] memory commandParams,
+        bytes32 nullifier
+    )
+        internal
+        override(EmailRecoveryManager)
```
JohnGuilding, line 260: Hmm do you know why this contract only requires `override(EmailRecoveryManager)`, but the production module in `clave-email-recovery` requires `override(EmailAccountRecovery, EmailRecoveryManager)`?

JohnGuilding (commented):
Thanks for putting this together. The code looks the same but noticed two inconsistencies:
1. override multiple inheritance
2. `deploymentTimestamp` in constructor 

I am concerned basing tests on commenting in/out code is error prone. Since this is going to production imminently, I'm in favour of building a higher level of confidence with the tests cases - Wdyt?

I wrote a non-exhuastive list of test cases that I was envisioning. Thinking here is we would cover every case and all tests would pass:

### Unit 

**Constructor** 
1. Modified constructor test to assert `deploymentTimestamp` was set correctly

**setTransactionInitiator**
1. Reverts if not owner
2. Sets non-zero address to true
3. Sets non-zero address to false after setting it to true (could be combined with 2.)
4. Sets zero address to true
5. Sets zero address to false after setting it to true (could be combined with 4.)

**acceptGuardian**
1. Reverts when transaction initiator is not set
2. Fails when exactly 6 * 30 days - 1 seconds has passed
3. Succeeds when transaction initiator is set
4. Succeeds when exactly 6 * 30 days has passed
5. Succeeds when exactly 6 * 30 days + 1 seconds has passed
6. Succeeds when zero address for transaction initiator is set to true

**processRecovery**
1. Reverts when transaction initiator is not set
2. Fails when exactly 6 * 30 days - 1 seconds has passed
3. Succeeds when transaction initiator is set
4. Succeeds when exactly 6 * 30 days has passed
5. Succeeds when exactly 6 * 30 days + 1 seconds has passed
6. Succeeds when zero address for transaction initiator is set to true

### Integration

1. Unapproved account cannot recover account (call initial acceptGuardian)
2. Can recover account with unapproved account(s) after 6 months (full recovery flow)
3. Can acceptGuardian or processRecovery with approved account -> permission is revoked for that account -> cannot processRecovery or acceptGuardian again
4. Can acceptGuardian or processRecovery with approved account -> permission is revoked for that account -> cannot processRecovery or acceptGuardian again -> permission is given to new account and recovery can continue
5. Any others you can think of

src/modules/EmailRecoveryModule.sol
aalimsahin: Added

JohnGuilding:
Thanks Alim. Resolved by https://github.com/zkemail/email-recovery/pull/82

PR #83, opened 2025-03-04
fix: optimizer enabled with 200 runs for foundry

Dhruv-2003:
# Introduction 

`EmailRecoveryUniversalFactory` has runtime size of 43,480 bytes, which is over contract size limit (24,576 bytes), resulting into the deployment failure

## Solution

Enabling optimizer during deployment reduces the size to 23,758 bytes with 200 runs solves the issue

## Note

Runs can be increased to 20_000 as well to further optimize if required

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -17,6 +17,9 @@ ignored_warnings_from = [
 threads = 1
 gas_limit = "18446744073709551615"
 memory_limit = 1844674407370955161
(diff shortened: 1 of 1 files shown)


JohnGuilding (approved):

JohnGuilding:
Fix for failing test has just been merged, so was safe to merge this

PR #84, opened 2025-03-15
feat: Abstracting proof verifier in account recovery 

Dhruv-2003:
## Introduction

The current account-recovery flow is specifically design for email based guardians, using email proofs to recover the accounts. 

## Solution 

The idea is to abstract the proof verfier in the account recovery flow. It will support different verifier like zk-email, zk-jwt, OAuth, zk-email.nr 

Mutliple guardian verifier implementations are supported for the guardians using the same module, i.e. guardians with different verifier can recover a single account

## Key changes

- Added `IGuardianVerifier` interface for implementation of verifiers based on different schemes 
- Removed the email-guardian logic from recovery modules

## Notes

- EmailAccountRecovery.sol contract was moved temporarily into this repo
- Module naming can be updated


src/EmailAccountRecoveryZKSync.sol
JohnGuilding, line 1: You can ignore zksync completely. Feel free to leave zksync files as they are or revert all changes (apart from minimal changes required for EmailRecoveryManagerZkSync to compile)

package.json
```diff
@@ -37,6 +37,7 @@
     "@safe-global/safe-contracts": "1.4.1-2",
     "@zk-email/contracts": "6.3.2",
     "@zk-email/ether-email-auth-contracts": "1.1.0",
+    "erc7579-implementation": "^0.3.1",
```
JohnGuilding: Why did you add this out of interest?

src/EmailAccountRecovery.sol
```diff
@@ -0,0 +1,217 @@
...
+    /// inactivates the controller (this contract).
+    /// @param recoveredAccount The address of the account to be recovered.
+    /// @return bool True if the account is already activated, false otherwise.
+    function isActivated(
+        address recoveredAccount
+    ) public view virtual returns (bool);
+
+    function acceptGuardian(
+        address guardian,
+        address recoveredAccount
```
JohnGuilding: `recoveredAccount` is confusing naming from us. Can rename to `account` everywhere it is used if you want

src/EmailAccountRecovery.sol
```diff
@@ -0,0 +1,217 @@
...
+    /// @param recoveryDataHash The hash of the recovery data
+    /// @param proofData The proof data for the guardian verifier.
+    function handleRecovery(
+        address guardian,
+        address recoveredAccount,
+        bytes32 accountSalt,
+        bytes32 recoveryDataHash,
+        IGuardianVerifier.ProofData memory proofData
+    ) external {
+        // TODO: Do we need to compute the guardian in handleRecovery as well, or just take the address as an input
```
JohnGuilding: The reason we compute things in these functions is to align it with the payload the relayer expects for calls related to the `EmailAuth` circuits. For these new contracts, it's safe to break from this interface and way of doing things.

So we can just pass the required arguments if they are sufficiently generic for all of the verifiers we want to support

src/EmailAccountRecovery.sol
```diff
@@ -0,0 +1,217 @@
...
+    /// uses a provided account address to be recovered, account salt, and
+    /// the hash of the encoded ERC1967Proxy creation code concatenated with the encoded guardian verifier implementation
+    /// address and the initialization call data. This ensures that the computed address is
+    /// deterministic and unique per account salt.
+    /// @param guardianVerifierImplementation The address of the guardian verifier implementation.
+    /// @param recoveredAccount The address of the account to be recovered.
+    /// @param accountSalt A bytes32 salt value used to ensure the uniqueness of the deployed proxy address.
+    /// @param verifierInitData The initialization data for the guardian verifier.
+    /// @return address The computed address.
+    function computeGuardianVerifierAddress(
```
JohnGuilding, line 66: Maybe we want to support non-upgradeable guardian verifiers, in which case this logic would change. Hmm let’s leave it for now, but will think more on this

src/EmailAccountRecovery.sol
```diff
@@ -0,0 +1,217 @@
...
+    /// @param guardianVerifierImplementation The address of the guardian verifier implementation.
+    /// @param recoveredAccount The address of the account to be recovered.
+    /// @param accountSalt A bytes32 salt value used to ensure the uniqueness of the deployed guardian verifier.
+    /// @param verifierInitData The initialization data for the guardian verifier.
+    /// @param proofData The proof data required for the guardian verifier.
+    /// the command in the given email auth message.
+    function handleAcceptance(
+        address guardianVerifierImplementation,
+        address recoveredAccount,
+        bytes32 accountSalt,
```
JohnGuilding, line 140: One small idea to simplify the interface is to pass the accountSalt as part of the proofData parameter. It doesn't matter too much for the verifiers we want to support for the grant scope as believe they would all use account code, but for additional non-zkemail zk verifiers, this would change

src/interfaces/IGuardianVerifier.sol
```diff
@@ -0,0 +1,72 @@
...
+     * Deployer is AccountRecovery module contract
+     *
+     * @param recoveredAccount Account to be recovered
+     * @param accountSalt Account salt
+     * @param initData Initialization data
+     */
+    function initialize(
+        address recoveredAccount,
+        bytes32 accountSalt,
+        bytes calldata initData
```
JohnGuilding, line 46: What use cases did you have in mind for this `initData` again?

src/interfaces/IGuardianVerifier.sol
```diff
@@ -0,0 +1,72 @@
...
+    /**
+     * @dev Verification logic of the proof
+     * @notice Reverts if the proof is invalid
+     *
+     * @param recoveredAccount Account to be recovered
+     * @param proof Proof data
+     * @return isVerified if the proof is valid
+     *
+     */
+    function verifyProofStrict(
```
JohnGuilding: Can you explain the rationale again for having `verifyProofStrict` and `verifyProof`

JohnGuilding (commented):
Thanks Dhruv. Let me get back to you tomorrow about where this should be merged. Some thoughts on those questions:

**Who should choose between different verifiers?**
The end user chooses the guardian type e.g. ethereum address, email, JWT. But the developer/client side code makes any lower level decisions e.g. circom or noir

**How should we choose between different verifiers?**
Not sure yet. When choosing between circom and noir, maybe a guardian is on a phone when handling acceptance but is on their computer when when handling recovery, in which case, would you want to dynamically choose between client side (noir) or server side (circom) proving?

Can keep this as an open question. For now lets keep it simple and pass something to indicate a guardian type in when adding a guardian. All of the guardian types can be supported by default

**How do we handle nullifiers/replay protection?**
haven't had enough time to think about this. Will keep you posted

src/EmailAccountRecovery.sol
JohnGuilding: Leaning towards non-upgradeable guardians

JohnGuilding: we were having similar conversations when designing the `EmailSigner`, and came to the conclusion we would remove upgradeability by default and only use it where it is needed (email signer would not be upgradeable, but the underlying verifier implementation is used would be)

For this PR, if you have time, would simplify and switch to non-upgradeable guardians

JohnGuilding: https://github.com/zkemail/email-tx-builder/blob/main/packages/contracts/src/EmailSigner.sol

JohnGuilding:
If you bump modulekit to latest that should resolve the CI failure

package.json
Dhruv-2003: This was not a dependency, but contracts had been used from erc7579 in the codebase, I was not sure why it wasn't already there.

Syntax highlighter were throwing errors if it wasn't a dependency.

If there is a another way around (maybe adding a remapping) , then we can go with that

src/interfaces/IGuardianVerifier.sol
Dhruv-2003: `verifyProofStrict` - Recommended to use when proof verification is done on-chain or when called from another contract. This avoids extra gas costs from handling failure cases manually.

`verifyProof` - Recommended to use when proof verification is done off-chain, saves gas cost. This allows clients to check errors without using extra gas.

src/interfaces/IGuardianVerifier.sol
Dhruv-2003: On top of my mind, it could be to setup access restriction based parameters, mostly taking reference from `EmailAuth.sol`

I am not sure about it's user in other cases

src/EmailAccountRecovery.sol
Dhruv-2003: Need to keep accountSalt separately as it's the salt to the create2 address of the new guardian-verifier as well, `computeGuardianVerifierAddress`. I just think it keep consistency in the functions if this arg is separate.

src/EmailAccountRecovery.sol
Dhruv-2003: I understand, but I would still want to use proxy-clone method for guardian verifiers, so that the deployment cost is minimal.

I can sure remove the upgradeability part here. Let me know if clones work according to you - https://docs.openzeppelin.com/contracts/4.x/api/proxy#Clones

src/EmailAccountRecoveryZKSync.sol
Dhruv-2003: Sure, I have done minimal changes to get it to compile, can remove it from PR if required

Dhruv-2003:
> If you bump modulekit to latest that should resolve the CI failure

After update, the CI failures are because of the incompleted tests for now, will work on them.

src/EmailAccountRecovery.sol
JohnGuilding: Yeah that's fair, Clones work

src/EmailAccountRecovery.sol
JohnGuilding: ah yeah true. Let's keep it as is.

We will look to make this interface more generic for our v2, which would include making the create2 values generic, or separating the guardian deployment and handleAcceptance calls, but that can be out of scope for this work.

src/interfaces/IGuardianVerifier.sol
JohnGuilding: Understood, in that case, would propose 2 changes:
1. Not against code duplication necessarily, but because the duplicate code is relatively large and complex, we could implement a design similar to OZ's [ECDSA library](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/ECDSA.sol) where we have single function for the core logic which returns error strings, then that can be wrapped in the function that is called onchain which reverts with the appropriate errors like [so](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/6015d7ef79172976b949be6e4b6c495593983c99/contracts/utils/cryptography/ECDSA.sol#L169-L179)
2. rename `verifyProof` -> `tryVerifyProof` & rename `verifyProofStrict` -> `verifyProof` 

> This avoids extra gas costs from handling failure cases manually.

Not sure I get this, can you elaborate? You mean gas costs from doing something like `_throwError`?

> saves gas cost. This allows clients to check errors without using extra gas.

`verifyProofStrict` is view so we wouldn't have to spend gas. The benefit is UX here? (which I'm in favour of)

src/EmailAccountRecoveryZKSync.sol
JohnGuilding: not needed, whatever you prefer

package.json
JohnGuilding: Yeah the remapping for it was pointing to it nested within another rhinestone dependency from memory.

Happy to leave it if it solved a remapping issue

Dhruv-2003: I'll confirm if that is the case

src/interfaces/IGuardianVerifier.sol
Dhruv-2003: > Not sure I get this, can you elaborate? You mean gas costs from doing something like _throwError?

yeah user wouldn't have to revert conditionally, but in an ideal scenario revert is always better from the verifier itself

Actually, now when I think more about it, It is not making much sense to return the error string instead of revert. Both offer similar UX and there is no gas benefit of `tryVerifyProof`

package.json
Dhruv-2003: Resolved this, no need to add this separately, altho I've updated the remapping, it was confusing the mapping for erc7579, can't have both `erc7579/` & `erc7579/interfaces`

src/EmailAccountRecovery.sol
Dhruv-2003: I checked this further post our conversation on the call, the current implementation with ERC1967 is in fact not upgradeable by default, it's just a proxy, it become upgradeable if you were to use `ERC1967Upgrade` or the implementation contract ( that being guardian verifier) follow an upgradeable pattern. So it's safe to leave it as such

DashaBochkar (approved):

src/interfaces/IGuardianVerifier.sol
Dhruv-2003: update the structure as per our discussion and handling the nullifier temporarily in the verifier as well, let me know your thoughts on it

Dhruv-2003:
I think this is ready for initial review, as the contract architecture is pretty much concrete & design is complete for now.

A few things to note & questions
- Contract and file naming are still with Email recovery module, should we update them or keep it ?
-  There are some changes my local prettier formatted a few files with, if there is a specific config I need to format this with, do share that
- Should we add the EIP7702 based tests in the same PR or move to a new PR for that, once we merge these base ones ?

JohnGuilding (approved):
> Contract and file naming are still with Email recovery module, should we update them or keep it ?

Keeping the same naming works if I understand this comment

> There are some changes my local prettier formatted a few files with, if there is a specific config I need to format this with, do share that

Happy to go with your formatting, we are looking into formatter configuration in some tech debt as should be formatting the same

> Should we add the EIP7702 based tests in the same PR or move to a new PR for that, once we merge these base ones ?

Let's get this PR merged and add later

JohnGuilding:
Approving this so we can get it in but we need to fix build in next PRs

PR #87, opened 2025-04-02
feat: Email guardian verifier added 

Dhruv-2003:
## Introduction 

The email guardian was tightly coupled in the recovery modules

## Solution 

The email guardian logic was moved into a new `EmailGuardianVerifier.sol` module that implements of the `IGuardianVerifier` . 

Verification logic for proof is migrated from the older EmailAccountRecovery-EmailAuth implementation. Everything related to just email verification is now handled in this verifier.

## Notes

- Certain config addresses are hardcoded in initialisation
- Command templates are added automatically from the command handler
- Further tests to be added, only unit tests remaining ??
- Remappings & foundry.toml were temporarily modified to work with this branch locally, if required the changes can be reversed before merging


Dhruv-2003:
Discussions until now: https://github.com/Dhruv-2003/email-recovery/pull/1

pnpm-lock.yaml
JohnGuilding, line 1: what is causing this change? Cannot see any diff in package.json

remappings.txt
```diff
@@ -24,3 +24,5 @@ solidity-stringutils/=node_modules/solidity-stringutils/
 @matterlabs/zksync-contracts/l2/contracts/=src/libraries/
 kernel/=node_modules/@zerodev/kernel/src/
 ExcessivelySafeCall/=node_modules/excessively-safe-call/src/
+
+src/=src/
```
JohnGuilding: What error were you getting that was avoided by this?

src/EmailGuardianVerifier.sol
```diff
@@ -0,0 +1,488 @@
...
+     * @param {account} The address of the account to be recovered.
+     * @param _accountSalt The salt used to derive the account address.
+     * @param initData The initialization data.
+     */
+    function initialize(
+        address /* account */,
+        bytes32 _accountSalt,
+        bytes calldata initData
+    ) public initializer {
+        // NOTE: Temporary bypass, remove after the upgrade
```
JohnGuilding, line 181: why is this bypass needed again?

src/EmailGuardianVerifier.sol
```diff
@@ -0,0 +1,488 @@
...
+
+            require(
+                account == _account,
+                InvalidAccountInCommand("acceptance", _account, account)
+            );
+
+            templateId = computeAcceptanceTemplateId(emailData.templateIdx);
+        }
+
+        require(emailData.isCodeExist == true, CodeDoesNotExist());
```
JohnGuilding: This should only be checked for handle acceptance part of the flow. From readme:

> - As described in the Subsection of "Invitation Code", for each email user, your application contract must ensure that the value of `isCodeExist` in the first email-auth message is true.

src/EmailGuardianVerifier.sol
```diff
@@ -0,0 +1,488 @@
...
+                revert InvalidCommand();
+            }
+        }
+
+        require(
+            verifier.verifyEmailProof(emailProof) == true,
+            InvalidEmailProof()
+        );
+
+        return (true);
```
JohnGuilding: nit: can drop brackets?

src/EmailGuardianVerifier.sol
```diff
@@ -0,0 +1,488 @@
...
+     * e.g. Recovery functions ( handleAcceptance & handleRecovery )
+     *
+     * @notice Nullifier based check is handled here
+     * @notice Timestamp check of when the proof was generated is handled here
+     * @notice Reverts if the proof is invalid
+     *
+     * @param account Account to be recovered
+     * @param proof Proof data
+     * proof.data: EmailData
+     * proof.publicInputs: [publicKeyHash, emailNullifier]
```
JohnGuilding: What is your thinking here behind naming public inputs? Does it say somewhere in our docs/code that these are the public inputs?

JohnGuilding (approved):

JohnGuilding:
Small merge conflict also

remappings.txt
Dhruv-2003: In the tests folder, it could't find the files under the `src` folder for some reason, this might be something related to my solidity extension.

pnpm-lock.yaml
Dhruv-2003: Didn't realise this was the case, I can undo it after checking this. Shouldn't have changed tho.

src/EmailGuardianVerifier.sol
Dhruv-2003: In the earlier version, Ownable.sol was used. As it turns out the registry needs the owner function, I had skipped in the first version, and ended up adding a basic owner set & get myself, just to keep it slim. 

Can move it to Ownable if that's preferred. I remember you guys mentioning something that this is only the case for `UserOverridableDKIMRegistry` and you are working on maybe updating this, that's why left the note.

src/EmailGuardianVerifier.sol
Dhruv-2003: So this was a common practice I followed by taking reference from `Email.Nr` circuits. In this case ( groth16 ) proofs, I couldn't find a doc saying this, but these parameters are anyways public on chain ( publicKeyHash for DKIM registry and nullifier in records )

Do you want to follow another approach ?

src/EmailGuardianVerifier.sol
Dhruv-2003: Thanks for the catch, updated

Dhruv-2003:
> Small merge conflict also

- Conflict resolved

- Resolved the issues noted in the comments
- Another small update here, `recoveryDataHash` is also parsed from the command template and verified, it's passed as as an argument in `EmailData` in proof.

remappings.txt
JohnGuilding: hmm yeah we'll re-examine our dependency resolution when building our solidity template repo cc @benceharomi 

Also does swapping to hardhat solidity extension change anything? can't remember if we checked before but also assuming you have no User settings that would override Workspace settings

src/EmailGuardianVerifier.sol
JohnGuilding: ah yeah I remember. Either using Ownable or keeping this works. I think the comment can be removed as we wouldn't upgrade anything for some time and not sure if we will do that update yet

src/EmailGuardianVerifier.sol
JohnGuilding: Happy with the approach, I just don't know exactly what the public inputs are for the EmailAuth circuits.

@zkfriendly what are the public inputs to the EmailAuth circuits?

zkfriendly: These are the public inputs for the [`EmailAuth` circom circuit](https://github.com/zkemail/email-tx-builder/blob/db9b113165c822c716ef6e72a573d9504428b60d/packages/circuits/src/email_auth_template.circom#L65):

```
struct EmailProof {
    string domainName; // Domain name of the sender's email
    bytes32 publicKeyHash; // Hash of the DKIM public key used in email/proof
    uint timestamp; // Timestamp of the email
    string maskedCommand; // Masked command of the email
    bytes32 emailNullifier; // Nullifier of the email to prevent its reuse.
    bytes32 accountSalt; // Create2 salt of the account
    bool isCodeExist; // Check if the account code is exist
    bytes proof; // ZK Proof of Email
}
```

JohnGuilding: Thanks @zkfriendly 



Clarifying question. So the full list of public outputs are:
```
    signal output domain_name[domain_field_len];
    signal output public_key_hash;
    signal output email_nullifier;
    signal output timestamp;
    signal output masked_command[command_field_len];
    signal output account_salt;
    signal output is_code_exist;
```

Which line up with the following values from the solidity struct:
```
    string domainName; // Domain name of the sender's email
    bytes32 publicKeyHash; // Hash of the DKIM public key used in email/proof
    uint timestamp; // Timestamp of the email
    string maskedCommand; // Masked command of the email
    bytes32 emailNullifier; // Nullifier of the email to prevent its reuse.
    bytes32 accountSalt; // Create2 salt of the account
    bool isCodeExist; // Check if the account code is exist
```

And then `bytes proof; // ZK Proof of Email` is or isn't a public output? The proof is separate right?

@Dhruv-2003 could you set the public outputs to match up with the above answer?

zkfriendly: > And then bytes proof; // ZK Proof of Email is or isn't a public output? The proof is separate right?

Yes, proof is separate. (proof + public inputs) is what is verified on-chain

Dhruv-2003: Thanks for the info, I do agree that's a better way.

But if I do that, it would add extra code for decoding the data from bytes32. I followed the current pattern as that's what the Verifier expects `EmailProof`. Should I go ahead with this ?

JohnGuilding: I'm not too opinionated right now. I think two options are:
1. Implement this change with all public outputs and add extra decoding logic
2. Drop public outputs and put those arguments somewhere else. Maybe `EmailData`? 

Will leave the decision up to you. Whatever you think is simplest?

Dhruv-2003: I wouldn't want to drop public inputs as it might be crucial for certain schemes ( noir ) to have.

`data` was an arbitrary field to add any extra info required, I will see what approach is a better one.

JohnGuilding: What do you think of an interface like this?

```solidity
function verifyProof(
    address account,
    uint256[2] memory a,
    uint256[2][2] memory b,
    uint256[2] memory c,
    uint256[] memory publicOutputs
) public returns (bool) {
    // decode public outputs
}
```

JohnGuilding: Or can have proof in a struct:

```solidity
function verifyProof(
    address account,
    Proof proof,
    uint256[] memory publicOutputs
) public returns (bool) {
    (
        uint256[2] memory a,
        uint256[2][2] memory b,
        uint256[2] memory c
    ) = abi.decode(proof, (uint256[2], uint256[2][2], uint256[2]));
    
    // decode public outputs
}
```

Dhruv-2003: Correct me if i am wrong, but this is only applicable on groth16 circuits, doesn't apply for noir and constrains the Zk Proof schemes this could support, and can you clarify why decoding at the guardian verifier level is required ?

Are there issues with the current Guardian Verifier interface ?
```
    /// @dev Proof data required proof verification
    struct ProofData {
        /// @dev Proof
        bytes proof;
        /// @dev Public inputs required for verification
        bytes32[] publicInputs;
        /// @dev Additional Data required for verification
        bytes data;
    }
```

Just for uniformity, it's `public inputs` for the circuits.

I can try going with the option 1 described above, that puts all the public inputs into the `bytes32[]` array and then decoding, but I feel it's an un-necessary extra step. Are there specific concerns with that design with having both data and publicInputs ?

JohnGuilding: > Are there specific concerns with that design with having both data and publicInputs ?

Nope. 

> Are there issues with the current Guardian Verifier interface ?

No would be happy with that if that's easier for compatibility with other verifiers. I do not feel confident I know non-groth16 verifiers. Can we keep `ProofData` as it is but then just make sure the public outputs are updated to reflect the above conversation?

> and can you clarify why decoding at the guardian verifier level is required ?

Maybe I'm not understanding the question fully, but we need to decode the values before calling the verifier? It is not required, I guess we could do it in the module, but that sounds more difficult. Let me know if I am misunderstanding the Q

Dhruv-2003: To clarify what I have been talking about, this is what the proof.data consist of
```
// Struct to hold proof data
    struct EmailData {
        // Template index for the command
        uint256 templateIdx;
        // Command parameters for the email
        bytes[] commandParams;
        // Skipped command prefix
        uint256 skippedCommandPrefix;
        // Domain name for the email
        string domainName;
        // Timestamp of the email
        uint256 timestamp;
        // Masked command for the email
        string maskedCommand;
        // Account salt used to derive the account address
        bytes32 accountSalt;
        // If the code exists
        bool isCodeExist;
        // If the email is for recovery
        bool isRecovery;
        // Recovery data hash ( only required for recovery verification )
        bytes32 recoveryDataHash;
    }
```

Few of these arguments are public inputs as well, as zk-friendly mentioned above, If we are to place all the field in publicInputs itself, I will have to pass them all as a bytes32 value, and then decode them separately to their respective type in the verifyProof function to match what you said. Qn is if the overhead of decoding is worth it for consistency ?

Let me know if it's still unclear

JohnGuilding: > Qn is if the overhead of decoding is worth it for consistency ?

If we could have a consistent verifier interface then would be in favour of the overhead

If we moved these values to `publicInputs` and then the interface was not-consistent with the Noir verifier, then would be in favour of keeping the initial interface. With the addition of ensuring the public inputs are consistent. E.g. I think we should avoid having only two public inputs listed like `[publicKeyHash, emailNullifier]`, when in fact there are more but in `EmailData`. In that case, we could consider removing publicInputs like so?

```solidity
    struct ProofData {
        /// @dev Proof
        bytes proof;
        /// @dev Additional Data required for verification
        bytes data;
    }
```

Let's jump on a call if this answer does not provide good enough clarification to resolve this comment

PR #88, opened 2025-04-02
feat: Jwt Guardian verifier added

Dhruv-2003:
## Introduction
Using jwt as a verification method for guardians was not possible

## Solution
The Jwt guardian is now added as JwtGuardianVerifier.sol module that implements of the IGuardianVerifier .

Verification logic is basic JWT verification which was taken as reference from [JwtVerifier.sol](https://github.com/zkemail/jwt-tx-builder/blob/main/packages/contracts/src/utils/JwtVerifier.sol)

## Notes
- Currently only the Jwt groth16 proof is being verified in the verification logic, needs to be updated on what else to verify. 
- Specific claims are not verified here for acceptance & recovery
- JwtRegistry verification is unclear and not implemented, to be discussed

### DIFF
--- a/src/JwtGuardianVerifier.sol
+++ b/src/JwtGuardianVerifier.sol
@@ -1,0 +1,258 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.25;
+
+import {IDKIMRegistry} from "@zk-email/contracts/DKIMRegistry.sol";
+
+import {IGuardianVerifier} from "./interfaces/IGuardianVerifier.sol";
+import {Initializable} from "@openzeppelin/contracts/proxy/utils/Initializable.sol";
+import {IVerifier, EmailProof} from "./interfaces/IJwtVerifier.sol";
+
+/**
+ * @title JwtGuardianVerifier
+ * @notice Provides a mechanism for guardian verification using jwt token proofs
+ * @dev The underlying IGuardianVerifier provides the interface for proof verification.
+ */
+contract JwtGuardianVerifier is IGuardianVerifier, Initializable {
+    // NOTE: Temporary bypass, remove after the upgrade
+    address _owner;
+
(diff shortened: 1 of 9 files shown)


src/JwtGuardianVerifier.sol
```diff
@@ -0,0 +1,281 @@
...
+     * proof.publicInputs: [publicKeyHash, emailNullifier]
+     * proof.proof: zk-SNARK proof
+     *
+     * NOTE: Specific claim is not yet verified
+     * NOTE: JWTRegistry check is disabled
+     * NOTE: Gas optimisation possible by only decoding the proof data once in this function rather in tryVerify as well
+     *
+     * @return isVerified if the proof is valid
+     */
+    // Lint Error: Cyclomatic complexity ( too many if else statements )
```
JohnGuilding: old comment about complexity

src/JwtGuardianVerifier.sol
```diff
@@ -0,0 +1,281 @@
...
+            publicKeyHash: proof.publicInputs[0],
+            timestamp: jwtData.timestamp,
+            maskedCommand: jwtData.maskedCommand,
+            emailNullifier: proof.publicInputs[1],
+            accountSalt: jwtData.accountSalt,
+            isCodeExist: jwtData.isCodeExist,
+            proof: proof.proof
+        });
+
+        // TODO: Handle JWT Registry check
```
JohnGuilding: yeah if it's possible to implement it lets do this. If it's not possible within the timeframe could you leave a comment explaining what changes are needed/error you're running into?

JohnGuilding (approved):

src/JwtGuardianVerifier.sol
Dhruv-2003: Implemented

PR #89, opened 2025-04-02
feat: Zkemail.Nr verifier guardian

Dhruv-2003:
## Introduction
Using zkemail noir circuits was not possible with guardians. With noir proofs, client side proving on mobile as well as other devices becomes possible, so it's necessary to have that.

## Solution
The ZkEmail.nr guardian is now added as EmailNrGuardianVerifier.sol module that implements of the IGuardianVerifier .

Verification logic is basic zkemail.nr proof verification which was taken as reference from noir solidity verifier [TestVerifier.sol](https://github.com/zkemail/zkemail.nr/blob/feat/solidity-verifier/js/sol/contracts/TestContract.sol)

## Notes
- Currently only the zkemail noir proof is being verified in the verification logic, other email related verification, like command template logic is not yet added.
- As the noir circuits are currently not built for verifying commands in the email, the same is reflected here, the proof verification is only for the email being a valid one, but not specific to acceptance and recovery. This is necessary to have in the actual flow.
- HonkVerifiers and other testing related are temporarily added in this repo, can be removed if exported from zk-email package

Tested with real noir proofs created locally using the noir prover.


src/EmailNrGuardianVerifier.sol
```diff
@@ -0,0 +1,251 @@
...
+     * @return isVerified if the proof is valid
+     */
+    function tryVerifyProof(
+        address account,
+        ProofData memory proof
+    ) public view returns (bool) {
+        // Parse the extra data
+        EmailData memory emailData = abi.decode(proof.data, (EmailData));
+
+        // TODO: How to handle dkimRegistry validation in case of noir circuits
```
JohnGuilding, line 220: hmm good question. I'm not 100% certain but I would hope we can use it in the same way as for non-noir flow. Was there any reason it could not be implemented here?

test-new/EmailNrGuardianVerifier/EmailNrGuardianVerifierBase.t.sol
```diff
@@ -0,0 +1,460 @@
...
+        emailProof.domainName = "gmail.com";
+        emailProof.timestamp = block.timestamp;
+        emailProof.maskedCommand = "";
+        emailProof.accountSalt = accountSalt;
+        emailProof.isCodeExist = true;
+
+        string memory proofFile = vm.readFile(
+            string.concat(
+                vm.projectRoot(),
+                "/test-new/EmailNrGuardianVerifier/recovery-proof.json"
```
JohnGuilding, line 392: This proof works and the integration test passes?

JohnGuilding: oh yeah remember your comment from TG you said it works - nice

test-new/EmailNrGuardianVerifier/EmailNrGuardianVerifierMock/EmailNrGuardianVerifier.t.sol
JohnGuilding, line 1: Why `OwnableValidatorRecovery_EmailNrGuardianVerifierMock_Test` and `OwnableValidatorRecovery_EmailNrGuardianVerifier_Test`?

test-new/EmailNrGuardianVerifier/EmailNrGuardianVerifierBase.t.sol
```diff
@@ -0,0 +1,460 @@
...
+        emailProof.domainName = "gmail.com";
+        emailProof.timestamp = block.timestamp;
+        emailProof.maskedCommand = "";
+        emailProof.accountSalt = accountSalt;
+        emailProof.isCodeExist = true;
+
+        string memory proofFile = vm.readFile(
+            string.concat(
+                vm.projectRoot(),
+                "/test-new/EmailNrGuardianVerifier/proof.json"
```
JohnGuilding: Which file is this I can't see it?

JohnGuilding (approved):

test-new/EmailNrGuardianVerifier/EmailNrGuardianVerifierBase.t.sol
Dhruv-2003: Oh this was removed, this whole function is no longer needed.

test-new/EmailNrGuardianVerifier/EmailNrGuardianVerifierMock/EmailNrGuardianVerifier.t.sol
Dhruv-2003: This is mockHonkVerifier based verification, so the proofs are mock ones as well, and thus the name.

src/EmailNrGuardianVerifier.sol
Dhruv-2003: There was no information on how DKIM registry comes into the flow in case of noir circuits. Another catch here is the the prover generates a different publicKey everytime, so it might require to set public key again. I'm not too aware on the internals of it as well in zkemail.nr.

More info would help, but I doubt that can be done under the timeline.

JohnGuilding: Understood. Could you mention this in the document for TODO `4. Collect blockers/issues with our stack into notes`

JohnGuilding: And let's leave it as is

test-new/EmailNrGuardianVerifier/EmailNrGuardianVerifierMock/EmailNrGuardianVerifier.t.sol
JohnGuilding: To rephrase my question, are we duplicating code unnecessarily here if one test is mocked and other test isn't? Should we just have the non-mocked tests

This question is based on the assumption that these two test files are basically the same apart from the fact one is mocked and one is not - let me know if this assumption is wrong

Only other difference I picked up was the setup is slightly different in one - 1 guardian rather than 3

Dhruv-2003: Oh, in that sense, I presume you are right, they are the exact same test, just the difference being mock proof and real one, in setup difference is `MockHonkVerifier` & `HonkVerifier` , in this case mock one can be removed

PR #90, opened 2025-04-04
test: abstracted verifier guardian e2e test added 

Dhruv-2003:
## Introduction
This PR adds the end to end test of using the abstracted verifier setup with multiple guardians like email, jwt & email.nr to recover one account.

It takes the reference from respective guardian verification and creates an end to end test for this situation that involves setting up this multi guardian verifiers, installing the module, and consequently the recovery-flow.

This is a forge based test, as it was the preferred way instead of the scripts method.

## Notes
- In the current setup, email.nr circuits use a real proof that was generated locally 
- jwt circuits use a mock groth16 verifier underneath, so the zk-proof verification is skipped but everything else in the setup remains same. This was the case because of not being able to setup the prover for generating the proof.
- email groth16 circuit similar to the jwt circuits uses a mock groth16 verifier, but other checks like command template are still performend.
- As the base branched including the verifiers were not yet merged to a single branch, I did so on my own, and thus all the extra changes noticed here. I will rebase this branch once all the other PRs are merged.

## Question 
- For this setup, I didn't go ahead with scripts as a forge test was possible and I wanted to mimic the same flow. Due to this integrating a real proof zk-email proof in the flow is a bit complex and requires setting up the local prover, yet to be done. Do you want me to work on this ?
- If scripts path is preferred than that's an extra step here, in that case I can use the relayer, but still a zk-email setup on a local testnet would be required. Do you want me to opt for this route instead of the current one ?


DashaBochkar (approved):

Dhruv-2003:
Moving the work to a new branch with PR #91

PR #91, opened 2025-04-08
test: abstracted verifier guardian e2e test added

Dhruv-2003:
## Introduction
This PR adds the end to end test of using the abstracted verifier setup with multiple guardians like email, jwt & email.nr to recover one account.

It takes the reference from respective guardian verification and creates an end to end test for this situation that involves setting up this multi guardian verifiers, installing the module, and consequently the recovery-flow.

This is a forge based test, as it was the preferred way instead of the scripts method.

## Notes
- In the current setup, email.nr & email circuits use a real proof that was generated locally 
- jwt circuits use a mock groth16 verifier underneath, so the zk-proof verification is skipped but everything else in the setup remains same. This was the case because of not being able to setup the prover for generating the proof.


Dhruv-2003:
## Question
- Can add the EIP7702 auth test in this flow itself if required, yet to take a look in terms of setting up the account locally for test. If priority to merge this PR, can do the EIP7702 auth separately.

test-new/e2e/GuardianVerifierBase.t.sol
```diff
@@ -0,0 +1,846 @@
...
+        address account,
+        address guardian,
+        bytes32 _recoveryDataHash,
+        address emailRecoveryModule,
+        bytes32 accountSalt
+    ) public returns (IGuardianVerifier.ProofData memory proofData) {
+        bytes32 nullifier = generateNewNullifier();
+
+        string
+            memory command = "Recver account 0x..123 using recovery hash 0x..123";
```
JohnGuilding: nit: small typo

test-new/e2e/GuardianVerifier.t.sol
```diff
@@ -0,0 +1,213 @@
...
+        acceptGuardian(
+            GuardianType.EmailGuardian,
+            emailGuardianVerifierImplementation,
+            account,
+            guardians[0],
+            emailRecoveryModuleAddress,
+            accountSalt1,
+            emailGuardianVerifierInitData
+        );
+        acceptGuardian(
```
JohnGuilding, line 47: Looks great. For completeness, could we also use the jwt guardian in this test? Could just be for acceptance flow for simplicity

JohnGuilding (approved):

test-new/e2e/GuardianVerifier.t.sol
Dhruv-2003: The only issue is not having a hosted prover, If it's available or you have old proofs that are valid, then maybe I can give it a shot.

A local prover is not an option for me, it's consuming extreme resources.

test-new/e2e/GuardianVerifierBase.t.sol
```diff
@@ -524,14 +576,26 @@ abstract contract OwnableValidatorRecovery_AbstractedRecoveryModule_Base is
...
+        if (guardianType == GuardianType.MockEmailGuardian) {
+            bytes32 nullifier = generateNewNullifier();
 
-        EmailProof memory emailProof = getEmailProof(proofFile);
+            emailProof = generateMockEmailProof(
+                command,
+                nullifier,
+                accountSalt
+            );
+        } else if (guardianType == GuardianType.EmailGuardian) {
```
JohnGuilding, line 589: ~~is this function (`getAcceptanceEmailProofData`) only called in `acceptGuardian`? Does that mean that `guardianType` is always `GuardianType.MockEmailGuardian` and we would never hit this check?~~ Never mind can see that it is called at beginning of `acceptGuardian`

test-new/e2e/GuardianVerifierBase.t.sol
```diff
@@ -427,6 +469,7 @@ abstract contract OwnableValidatorRecovery_AbstractedRecoveryModule_Base is
     ) public {
         IGuardianVerifier.ProofData
             memory proofData = getAcceptanceEmailProofData(
```
JohnGuilding, line 471: Why do we call this here and also on line 494?

test-new/e2e/GuardianVerifierBase.t.sol
```diff
@@ -612,6 +676,7 @@ abstract contract OwnableValidatorRecovery_AbstractedRecoveryModule_Base is
     ) public {
         IGuardianVerifier.ProofData
             memory proofData = getRecoveryEmailProofData(
```
JohnGuilding, line 678: same comment, why do we call this here and on line 704?

JohnGuilding (approved):

test-new/e2e/GuardianVerifierBase.t.sol
Dhruv-2003: Actually I just noticed, I don't need to do call that again, I was dealing guardianType case by case so slipped up.

test-new/e2e/GuardianVerifierBase.t.sol
Dhruv-2003: https://github.com/zkemail/email-recovery/pull/91/files#r2037822289

test-new/e2e/GuardianVerifierBase.t.sol
JohnGuilding: Could you remove the duplicate call in the next PR? Happy to merge this now as only a small comment

Dhruv-2003: Ahh! I did locally, forgot to push 🥲

will do in the next one

PR #96, opened 2025-04-11
chore: cleanup & fix build issue 

Dhruv-2003:
## Introduction 

Build fails as the `forge-test-multi-account` errors out with test folders not building properly

## Solution 

- Old tests removed 
- Old scripts removed 
- Script updated with new contracts 
- New e2e and integration tests added with Guardian Verifiers under `test`
- fix : duplicate call [issue](https://github.com/zkemail/email-recovery/pull/91#discussion_r2038365847)


src/JwtGuardianVerifier.sol
```diff
@@ -226,8 +226,6 @@ contract JwtGuardianVerifier is IGuardianVerifier, Initializable {
         // Parse the extra data
         JwtData memory jwtData = abi.decode(proof.data, (JwtData));
 
-        require(jwtData.isCodeExist == true, CodeDoesNotExist());
```
JohnGuilding, line 229: @zkfriendly with the jwt verifier, what should we do with this check? For example, in the email recovery EmailAuth flow, we assert `emailData.isCodeExist == true` on the acceptance part of the flow, but not in the handle recovery part of the flow

Also related to the question I raised on your fixture PR

test/Base.t.sol
```diff
@@ -202,303 +150,88 @@ abstract contract BaseTest is RhinestoneModuleKit, Test {
...
-
-        bytes32 nullifier = generateNewNullifier();
-
-        bytes32 accountSalt;
-        if (optionalAccountSalt == bytes32(0)) {
-            accountSalt = getAccountSaltForGuardian(account, guardian);
-        } else {
-            accountSalt = optionalAccountSalt;
-        }
+        /// Prepare the initData with the bootstrap contract call along with the, can go with the generic one
```
JohnGuilding: Did this sentence get cut short here?

test/e2e/GuardianVerifier.t.sol
```diff
@@ -78,6 +78,12 @@ contract OwnableValidatorRecovery_AbstractedRecoveryModule_Test is
     // End to end test
     function test_Recover_RotatesOwnerSuccessfully() public {
         skipIfCommandHandlerType(CommandHandlerType.SafeRecoveryCommandHandler);
+        if (
+            accountAddress1 !=
```
JohnGuilding: This check is hard to understand at first glance, could `0x0929E9d9bC617836B650Ee4E419f52D31831C806` be assigned to a variable? And assume we are just enforcing that the expected address is correct? Should we revert instead of skip - or let me know if I am misinterpreting the check

JohnGuilding (approved):
Sweet, thanks for wrapping up the build and test changes. Only small comments

src/JwtGuardianVerifier.sol
Dhruv-2003: As far as I understand, this was a check for account code in the command, which doesn't exist here. The args are still passed as the jwt-verifier expects similars args as email-verifier.

test/e2e/GuardianVerifier.t.sol
Dhruv-2003: Actually I have added this check, as the only account this will pass is the one in argument, it's due to the because of the specific zk-email proof we generated and use to verify.

I can perhaps assign it to a variable if that's required.

test/Base.t.sol
Dhruv-2003: Oh yeah, seems like a mistake while adding the comment, will update with any other changes we'll have here

src/JwtGuardianVerifier.sol
JohnGuilding: Yeah wondering if we should ignore it or not in this case

test/e2e/GuardianVerifier.t.sol
JohnGuilding: Does it need to be a check? Can we just assume that the test is set up for a specific account? I'm not sure what the check is preventing if the test setup always uses this account

src/JwtGuardianVerifier.sol
JohnGuilding: Will see what zkfriendly says, if we don't have a concrete answer on whether the check should be here or not, we can merge the pr

test/e2e/GuardianVerifier.t.sol
Dhruv-2003: yeah, we don't need to check for that, we have the same account salt so it'll be the same address in the setup

Dhruv-2003: Added a note just for future reference

Dhruv-2003: Actually just remembered the CI, also runs test with other type of accounts like safe, etc. which fails the test as they will generate a different address even with the same salt.

Should I add it back in, or add an account type skip check ?

JohnGuilding: > Should I add it back in, or add an account type skip check ?

Right, yeah let's add it back in and leave a comment explaining why it is needed (also might as well put the address in a variable for readability)

Dhruv-2003: Done!

src/JwtGuardianVerifier.sol
zkfriendly: > with the jwt verifier, what should we do with this check? 
 
 I believe it is safe to remove this check here.  But I don't have full context on the jwt work.
 
> For example, in the email recovery EmailAuth flow, we assert emailData.isCodeExist == true on the acceptance part of the flow, but not in the handle recovery part of the flow

Thanks for sharing this. I now see an interesting point about this check. By including the accountCode in the email during the handle acceptance flow and verifying it on-chain, we have proof that the user knows the accountCode—and this will be the same accountCode used to accept the guardian request.

Removing this check allows a malicious relayer to tell the user that account code x will be used, while secretly using account code y behind the scenes. Later, the relayer can withhold y and stop sharing it with the user, effectively preventing them from ever being able to sign again.

JohnGuilding: We've discussed this in [another PR](https://github.com/zkemail/email-tx-builder/pull/127#discussion_r2045109640) and decided we need to think more about this check. This PR is good to go

PR #97, opened 2025-04-17
Feat/scaffold v2 recovery module

JohnGuilding:
This PR add scaffolds our V2 recovery module. It contains the functionality to install a module the module, vote on recovery attempts, and execute recovery. To start, we just have an ECDSA-based guardian:
* Scaffold the v2 recovery module
* Use with ECDSA guardian
* add an integration test

Can verify the test passes by running locally - need to re-add CI to this branch to run tests in CI

The PR is already 800 LOC so do not want to add too much more to make reviewing easier. Thinking we can add unit tests and additional functionality in subsequent PRs

Future improvements:
* Unit test coverage
* CI 
* Additional functionality
* Use factory to deploy guardian in test


src/RecoveryModule.sol
```diff
@@ -0,0 +1,328 @@
...
+        }
+        _;
+    }
+
+    /**
+     * @dev This function is called by the smart account during installation of the validator
+     * @param data arbitrary data that may be required on the validator during `onInstall`
+     * initialization
+     *
+     * MUST revert on error (i.e. if validator is already enabled)
```
zkfriendly, line 49: In this implementation there are no checks for this.

src/RecoveryModule.sol
```diff
@@ -0,0 +1,328 @@
...
+        uint256 guardianCount = guardians.length;
+        if (threshold == 0) {
+            revert ThresholdCannotBeZero();
+        }
+        RecoveryConfig storage recoveryConfig = recoveryConfigs[account][
+            validator
+        ];
+        for (uint256 i = 0; i < guardianCount; i++) {
+            address guardian = guardians[i];
+            if (guardian == address(0) || guardian == account) {
```
zkfriendly, line 69: Do we need to check that addresses in `guardians` list are unique as well? or is the assumption that they are unique ?

src/RecoveryModule.sol
```diff
@@ -0,0 +1,328 @@
...
+
+    /**
+     * @notice Initiates a recovery for a given account
+     * @dev called once per guardian
+     * @param account The account
+     * @param validator The validator to recover
+     * @param signature Signature/proof of guardian authentication
+     * @param hash The recovery calldata hash
+     */
+    function approveRecovery(
```
zkfriendly, line 184: I don't see any checks to make sure `guardian` in the `approveRecovery` is actually a `guardian` for the given `account` and `validator`. Is this expected or do we need extra checks here?

src/interfaces/IGuardian.sol
zkfriendly, line 1: curious to know your thoughts on using a standard interface like 1271 or [`IERC7913SignatureVerifier` from OZ](https://github.com/OpenZeppelin/openzeppelin-community-contracts/blob/e46e538e4fc7ce13986f1547e42f1b0eb4a81c8a/contracts/interfaces/IERC7913.sol#L16) since the arguments and expected behavior is essentially the same .  This is also fine too imo.

zkfriendly (commented):
Thanks, looks good. a few clarifying comments

JohnGuilding:
@zkfriendly @benceharomi holding off on these PR changes as we might be pivoting our approach to recovery V2

JohnGuilding:
Closed as no longer building our own recovery module

PR #99, opened 2025-04-21
Feat/email auth migration contracts

JohnGuilding:
This PR starts the migration of our email recovery logic from email-tx-builder. We need to migrate the recovery contracts, the relayer, and the zksync logic/contracts. This PR makes the following changes
* Migrate EmailAccountRecovery
* Migrate relevant tests
* Delete ZK Sync related contracts in email-recovery repo
* Not related to migration but deletes some useless scripts

This PR partly solves the following [linear issue](https://linear.app/zk-email/issue/SOL-6/migrate-email-recovery-branch-in-email-tx-builer-to-email-recovery)

Future work out of scope of this PR:
1. Get integration test working
2. Migrate relayer
3. Migrate ZK Sync contracts to this repo or it's own repo


test/unit/EmailAccountRecovery/EmailAccountRecovery_handleAcceptance.t.sol
```diff
@@ -0,0 +1,367 @@
...
+        emailAuthMsg.proof.accountSalt = 0x0;
+
+        vm.mockCall(
+            address(recoveryController.emailAuthImplementationAddr()),
+            abi.encodeWithSelector(EmailAuth.authEmail.selector, emailAuthMsg),
+            abi.encode(0x0)
+        );
+
+        vm.startPrank(someRelayer);
+        vm.expectRevert(bytes("guardian status must be REQUESTED"));
```
zkfriendly: This test is failing in this repo due to revert message being slightly different. The following should fix it:

```suggestion
        vm.expectRevert(bytes("status must be REQUESTED"));
```

zkfriendly (changes requested):
Looks good. One unit test is failing due to slightly different revert message.

test/unit/EmailAccountRecovery/EmailAccountRecovery_handleAcceptance.t.sol
JohnGuilding: Ah yes my bad, I rebased onto main to use the new pre-commit hooks and didn't run my tests after resolving the errors. This one was a max length linting error

JohnGuilding: fixed

zkfriendly (approved):

.solhintignore
```diff
@@ -1 +1,2 @@
-node_modules/
\ No newline at end of file
+node_modules/
+src/EmailAccountRecovery.sol
```
benceharomi, line 2: Why do we need to ignore this file?

benceharomi (approved):
Looks good. CI PR has been merged recently so we might want to trigger it to run on this PR too.

JohnGuilding:
Keeping this PR open but in draft until we resolve the conversation around relayer and circuits consolidation

.solhintignore
JohnGuilding: There are lots of linting and formatting errors with that file, but it is a core file that has been audited. So if we fixed all the issues, then we would have unaudited production code in this repo. The changes themselves should not introduce any new bugs, but imo this is not something we should be doing if we're taking security seriously. Let me know if you disagree

benceharomi: How do we track what is audited and what is not?

JohnGuilding:
Decided we will merge this

PR #103, opened 2025-06-18
fix: unsafe calldata access

benceharomi:
## What is this PR doing?

Unsafe Calldata Access in Function Selector Extraction fixed:
* assembly removed
* bounds checking using Solidity's native slice operator

## How Has This Been Tested?

Automated unit tests have been updated.

Tests added for cases when:
* calldata is empty
* calldata is less than 4 bytes

## Checklist:

- [x] I have performed a self-review of my own code
- [x] I have posted a link to the PR in the group chat
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have made corresponding changes to the documentation if required


src/modules/UniversalEmailRecoveryModule.sol
```diff
@@ -288,12 +288,16 @@ contract UniversalEmailRecoveryModule is
...
-            selector := mload(add(recoveryCalldata, 32))
+        // The calldata must be at least 4 bytes to contain a function selector
+        // This check prevents an out-of-bounds read
+        if (recoveryCalldata.length < 4) {
+            revert InvalidSelector(bytes4(0));
         }
 
+        // Safely extract the selector by casting the first 4 bytes of the calldata
+        // This is a safe operation due to the length check above
+        bytes4 selector = bytes4(recoveryCalldata);
```
zkfriendly: alternatively we could use the following code and avoid a custom error and a explicit length check:

```solidity
// In the ABI encoding of (address, bytes), the actual bytes content starts after:
// - 32 bytes for the dynamic array length
// - 32 bytes for the address
// - 32 bytes for the length of the bytes
// Then we want the first 4 bytes of the actual content (the function selector)
bytes4 selector = bytes4(recoveryData[96:100]);
 ```
 
 lmk what you think

**Review by zkfriendly — CHANGES_REQUESTED** [review]
thanks, looks good. Left one suggestion regarding function selector.

File: src/modules/UniversalEmailRecoveryModule.sol (hunk 1 above)
Comment (benceharomi) [reply src/modules/UniversalEmailRecoveryModule.sol]: I like it! To be a bit more precise:
```Solidity
// ABI encoding of (address validator, bytes calldata)
// - [32 bytes: address]
// - [32 bytes: offset to calldata]
// - [32 bytes: length of calldata]
// - [ x bytes: calldata]
//
// calldata starts at index 96 (32 + 32 + 32)
// first 4 bytes of calldata is the selector (96:100)
bytes4 selector = bytes4(recoveryData[96:100]);
```

zkfriendly (approved):

ernestognw (approved):

test/unit/modules/UniversalEmailRecoveryModule/recover.t.sol
```diff
@@ -38,14 +38,43 @@ contract UniversalEmailRecoveryModule_recover_Test is UnitBase {
         emailRecoveryModule.exposed_recover(accountAddress1, invalidCalldata);
     }
 
+    function test_Recover_RevertWhen_EmptyCalldata() public {
+        bytes memory emptyCalldata = bytes("");
+        bytes memory invalidCalldata = abi.encode(accountAddress1, emptyCalldata);
+
+        vm.startPrank(emailRecoveryModuleAddress);
+        // This should revert due to unsafe memory access when reading recoveryData[96:100]
```
ernestognw:
```suggestion
        // This should revert due to unsafe calldata access when reading recoveryData[96:100]
```

test/unit/modules/UniversalEmailRecoveryModule/recover.t.sol
```diff
@@ -38,14 +38,43 @@ contract UniversalEmailRecoveryModule_recover_Test is UnitBase {
...
+        vm.expectRevert();
+        emailRecoveryModule.exposed_recover(accountAddress1, invalidCalldata);
+    }
+
+    function test_Recover_RevertWhen_CalldataLessThan4Bytes() public {
+        // Create a scenario that triggers unsafe memory access
```
ernestognw:
```suggestion
        // Create a scenario that triggers unsafe calldata access
```

test/unit/modules/UniversalEmailRecoveryModule/recover.t.sol
```diff
@@ -38,14 +38,43 @@ contract UniversalEmailRecoveryModule_recover_Test is UnitBase {
...
+        // Create a bytes array with only 2 bytes (less than the 4 bytes needed for a selector)
+        bytes memory shortCalldata = new bytes(2);
+        shortCalldata[0] = 0x12;
+        shortCalldata[1] = 0x34;
+
+        // Encode this into recovery data format
+        bytes memory recoveryData = abi.encode(accountAddress1, shortCalldata);
+
+        vm.startPrank(emailRecoveryModuleAddress);
+        // This should revert due to unsafe memory access when reading recoveryData[96:100]
```
ernestognw:
```suggestion
        // This should revert due to unsafe calldata access when reading recoveryData[96:100]
```

ernestognw (commented):
just some nits for correctness, in reality the issue is not accessing memory out of bounds but rather an unspecified section of the `msg.data` sent in the call to the contract

